PR #265, opened 2019-01-28
Update deps, eslint settings, and fix some warnings that were happening

TreTuna:
Removed running tests on git hooks until we update them as they were hanging due to an async issue.

Update eslint settings to stop it from yelling at us about not knowing what version of React we were running.

Fix deprecated `scrollable` prop issue in `Tabs` component and use `variant` prop instead

Update all dependencies.


algernon:
This all looks good except for one thing:

screenshot from 2019-01-28 08-29-42

That menubar... it was hidden before. We do call `Menu.setApplicationMenu(null)`, but it seems to have no effect on Linux (despite the docs). I think we should postpone the Electron upgrade until this can be fixed.

algernon:
Thanks!

PR #266, opened 2019-01-29
Break out DeviceMenu, BoardMenu, BoardMenuItem

TreTuna:
This is the first PR of the process for pulling the smaller components out of the render method in `App.js` and into their own for separation and testing.


TreTuna:
Extracted a lot more.... note Commits

PR #358, opened 2019-03-18
Reorganize Editor components and pull out Dialog components to their own files

TreTuna:


TreTuna:
Bah, yeah, let me clean this up. I see a couple things I wanted to do. Had this sitting in my back pocket from the other night and didn't realize where it was exactly.

TreTuna:
I advanced this to what I meant to do originally. Not just breaking out those components, but grouping all Editor components into its folder and using an `index.js` file to export the default component for easy imports elsewhere.

PR #370, opened 2019-03-29
Add electron-window-state

TreTuna:
Add electron-window-state
Set default window size to 1200x900
Set to save window size and position on exit and restore on startup

Solves: #368


TreTuna:
Will do!

PR #392, opened 2019-05-24
(#381) enhancement: use a nice colour picker

0957758592:
Used [`react-color`](https://casesandberg.github.io/react-color/) to implement new UI

**Issue** [#381 ](https://github.com/keyboardio/Chrysalis/issues/381)


algernon:
Can you amend the commit so it has a proper `Signed-off-by` line? (see [here](https://github.com/keyboardio/Chrysalis/pull/392/checks?check_run_id=135623419) for how to do that)

Thanks!

package.json
```diff
@@ -82,7 +82,7 @@
     "babel-loader": "^8.0.2",
     "electron": "4.1.1",
     "electron-builder": "^20.39.0",
-    "electron-webpack": "^2.0.1",
+    "electron-webpack": "^2.6.2",
```
algernon: Can you drop this change? Upgrading electron-webpack breaks our prod builds (see #365).

src/renderer/screens/ColorPalette/ColorButton.js
```diff
@@ -0,0 +1,85 @@
+/**
```
algernon: Can you add copyright headers for the new files? You can put yourself as copyright holder, don't need to put the same info there as in the rest of the files (but you can, if you're ok with having your copyright assigned to Keyboardio).

src/renderer/screens/ColorPalette/ColorButtonsArea.js
```diff
@@ -0,0 +1,101 @@
+/**
```
algernon: Same here, we have copyright in every file (as far as I remember), new ones should have them as well. You can copy & paste from any of the existing files and change the copyright holder to your name (or leave it at keyboardio, if you're ok with that).

src/renderer/screens/ColorPalette/PickerColorButton.js
```diff
@@ -0,0 +1,127 @@
+/**
```
algernon: Same: copyright headers.

src/renderer/screens/ColorPalette/index.js
```diff
@@ -0,0 +1,167 @@
+/**
```
algernon: Same as above. Sorry!

yarn.lock
```diff
@@ -36,7 +36,7 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/core@7.4.0":
+"@babel/core@7.4.0", "@babel/core@>=7.1.0", "@babel/core@^7.1.0", "@babel/core@^7.2.2":
```
algernon: I haven't tested if this breaks prod builds yet, but any `package.json` or `yarn.lock` changes need to be carefully wetted. So unless this is absolutely required for the functionality of the new color picker, please reset `yarn.lock` to the version on current master.

algernon:
I'll have a look at how it looks in Chrysalis too, so I can see how it actually looks, too.

algernon:
I checked the PR in practice too now using a Model 01, and a few things don't seem to work:

- The palette button doesn't appear to have neither hover, nor press animations like other floating action buttons. I suppose this is because `PickerColorButton` is a custom implementation. Could it be re-implemented as a [Floating Action Button](https://material-ui.com/components/buttons/#floating-action-buttons)?
- I can change palette colors fine, but I don't seem to be able to assign new colors to keys on the keymap. Previously, if I had a color selected, and clicked on a key, the key changed color. Now it doesn't, and there's nothing on the dev console either. Having a key selected, and selecting another color from the palette doesn't change anything, either.

The new picker does look nice, though!

0957758592:
> I checked the PR in practice too now using a Model 01, and a few things don't seem to work:
> 
> * The palette button doesn't appear to have neither hover, nor press animations like other floating action buttons. I suppose this is because `PickerColorButton` is a custom implementation. Could it be re-implemented as a [Floating Action Button](https://material-ui.com/components/buttons/#floating-action-buttons)?

I will check it.

> * I can change palette colors fine, but I don't seem to be able to assign new colors to keys on the keymap. Previously, if I had a color selected, and clicked on a key, the key changed color. Now it doesn't, and there's nothing on the dev console either. Having a key selected, and selecting another color from the palette doesn't change anything, either.

You can use `CTRL` or `SHIFT` to set the color for key after you have choose the color.
(Sorry, I'm a little bit mix it up with issue [#382](https://github.com/keyboardio/Chrysalis/issues/382) )
Will change and update it!

> 
> The new picker does look nice, though!
Thanks!

algernon:
> You can use `CTRL` or `SHIFT` to set the color for key after you have choose the color.
> (Sorry, I'm a little bit mix it up with issue [#382](https://github.com/keyboardio/Chrysalis/issues/382) )
> Will change and update it!

Ah, I see. I think it'd be best if we stuck with the old behaviour for now, and tackle #382 in a different PR. Makes it easier to review and merge things.

0957758592:
> > You can use `CTRL` or `SHIFT` to set the color for key after you have choose the color.
> > (Sorry, I'm a little bit mix it up with issue [#382](https://github.com/keyboardio/Chrysalis/issues/382) )
> > Will change and update it!
> 
> Ah, I see. I think it'd be best if we stuck with the old behaviour for now, and tackle #382 in a different PR. Makes it easier to review and merge things.

Okay! Sure!

src/renderer/screens/Editor/Editor.js
```diff
@@ -627,7 +624,7 @@ class Editor extends React.Component {
               </Tooltip>
               <Tooltip title={i18n.editor.clearLayer}>
                 <IconButton disabled={isReadOnly} onClick={this.confirmClear}>
-                  <LayersClearIcon />
+                  <HighlightOffIcon />
```
TreTuna: Same as above, the previous icon was decided upon last week.

src/renderer/screens/Editor/Editor.js
```diff
@@ -76,9 +76,6 @@ const styles = theme => ({
   layerItem: {
     paddingLeft: theme.spacing.unit * 4
   },
-  layerSelect: {
```
TreTuna: This entire file needs to be reverted. This is what was moving the layer picker to the right side. 

All of the changes in this file are reverting all of @algernon's changes from last week that had been agreed upon. Please see: https://github.com/keyboardio/Chrysalis/pull/389

src/renderer/screens/ColorPalette/ColorButton.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const styleInFocus = {
+    background: `rgb(${color.r}, ${color.g}, ${color.b})`,
+    boxShadow: !isWhiteColor
+      ? `0px 0px 26px 4px rgb(${color.r}, ${color.g}, ${color.b})`
+      : `0px 0px 26px 4px rgb(155, 155, 155)`
+  };
+
+  return (
+    <div className={classes.root}>
```
TreTuna: All this div and class are doing is adding 5px of padding, can we just add that padding to the button and get rid of this?

src/renderer/screens/ColorPalette/ColorButton.js
```diff
@@ -0,0 +1,102 @@
...
+      <div
```
TreTuna: Can we use an actual button for this instead of a div styled as a button, please? This is for accessibility reasons. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

src/renderer/screens/ColorPalette/ColorButtonsArea.js
```diff
@@ -0,0 +1,118 @@
...
+    setIsFocus,
+    palette,
+    disabled
+  } = props;
+
+  /**
+   * This is Hook that lets add React state "colorButtonsAmount" to functional components
+   * @param {array} [state] Array with eight color elements
+   */
+  const [colorButtonsAmount] = useState(
```
TreTuna: Why are we using `useState` here for something that never changes? This seems like it should be a static variable.

src/renderer/screens/ColorPalette/ColorButtonsArea.js
```diff
@@ -0,0 +1,118 @@
...
+    const color = {
+      r: colorFocusButton.r,
+      g: colorFocusButton.g,
+      b: colorFocusButton.b,
+      rgb: `rgb(${colorFocusButton.r}, ${colorFocusButton.g}, ${
+        colorFocusButton.b
+      })`
+    };
+    if (indexFocusButton < colorButtonsArrayLength && panelNumber === 0) {
+      colorButtonsAmount.splice(indexFocusButton, 1, { ...color });
```
TreTuna: Wait, if colorButtonsAmount is a factor of state, why are we using `splice` here and not `setColorButtonsAmount` and `slice`? This is directly mutating that state and I don't recommend going down that route.

src/renderer/screens/ColorPalette/ColorButtonsArea.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+    if (indexFocusButton >= colorButtonsArrayLength && panelNumber === 8) {
+      colorButtonsAmount.splice(indexFocusButton - colorButtonsArrayLength, 1, {
```
TreTuna: Here as well. We should not directly mutate state?

TreTuna:
A couple things on the file structure.

The entire `ColorPalette` folder should be under the `components` folder as it is not a full-screen component. It is just one component on the `editor` screen. 

Any `index.js` files should be a simple `import/export` file. The actual code for the component should be under a file with that component's name. In this case, please move the code currently in `index.js` to a file named `ColorPallete.js` and then the following should be in the `index.js` file:

```javascript
import ColorPalette from "./ColorPalette";

export default ColorPalette;

```

0957758592:
> ile with that component's



> A couple things on the file structure.
> 
> The entire `ColorPalette` folder should be under the `components` folder as it is not a full-screen component. It is just one component on the `editor` screen.
> 
> Any `index.js` files should be a simple `import/export` file. The actual code for the component should be under a file with that component's name. In this case, please move the code currently in `index.js` to a file named `ColorPallete.js` and then the following should be in the `index.js` file:
> 
> ```js
> import ColorPalette from "./ColorPalette";
> 
> export default ColorPalette;
> ```

Okay. Thank you!
Will be updated!

src/renderer/screens/ColorPalette/PickerColorButton.js
```diff
@@ -0,0 +1,144 @@
...
+    position: "relative",
+    margin: "0 20px",
+    [theme.breakpoints.down("sm")]: {
+      marginLeft: 10,
+      marginRight: 0
+    }
+  },
+  swatch: {
+    padding: 5,
+    width: 40,
```
TreTuna: Could we enlarge this to 60, please? See larger comment about resizing this entire palette.

src/renderer/screens/ColorPalette/PickerColorButton.js
```diff
@@ -0,0 +1,144 @@
...
+    height: 40,
```
TreTuna: Could we enlarge this to 60, please? See larger comment about resizing this entire palette.

src/renderer/screens/ColorPalette/index.js
```diff
@@ -0,0 +1,184 @@
...
+    left: 0,
+    right: 0,
+    bottom: 0,
+    height: 150,
+    justifyContent: "center",
+    alignItems: "center"
+  },
+  palette: {
+    display: "flex",
+    height: 70,
```
TreTuna: Could we enlarge this to 80, please? See larger comment about resizing this entire palette.

TreTuna:
I'd like to enlarge this entire palette just a smidge. Here's are comparing screenshots:

Current implementation:
image

Slightly enlarged palette and picker button. 
image

It gives the picker a little more emphasis and a much better ratio with the squared swatches, especially as smaller screen widths.

I've made code comments on the 3 lines needed for this simple change.

0957758592:
> I'd like to enlarge this entire palette just a smidge. Here's are comparing screenshots:
> 
> Current implementation:
> image
> 
> Slightly enlarged palette and picker button.
> image
> 
> It gives the picker a little more emphasis and a much better ratio with the squared swatches, especially as smaller screen widths.
> 
> I've made code comments on the 3 lines needed for this simple change.

Sure! Yes!

src/renderer/screens/ColorPalette/PickerColorButton.js
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    <div className={classes.root}>
+      <div
+        className={classes.swatch}
+        onClick={handleClick}
+        style={disabled ? styleDisabled : null}
+      />
+      {displayColorPicker ? (
+        <div className={classes.popover}>
+          <div className={classes.cover} onClick={handleClose} />
```
TreTuna: Curious on this... Due to this div, it causes the picker to close whenever anything on the lower 150px of the screen, but only that area, full-width x 150px high. Is this supposed to support clicking outside of the picker anywhere on the entire screen? It seems like an odd distinction and bit of UX distraction to me as it sits.

src/renderer/components/ColorPalette/ColorPalette.js
```diff
@@ -0,0 +1,164 @@
...
+      width: 470
+    }
+  }
+});
+
+/**
+ * Use to reduce the amount of code
+ * @param {object} color Object with keys that defining colors using the Red-green-blue-alpha (RGBA) model
+ */
+const setColorTamplate = color => ({
```
TreTuna: We have a function with this name in two different places and they do two different things. This one returns an object with `{r, g, b, rgb}` and the other returns a string with a css function `rgb(r, g, b)`. We should be able to reuse this one and where the other is used, just call this and take the `rgb` property from it. This way we could move this function to our `utils` folder and then imported into these components, this way it can be reused throughout the application.

src/renderer/components/ColorPalette/ColorButton.js
```diff
@@ -0,0 +1,113 @@
...
+  minHeight: value,
+  maxWidth: value,
+  maxHeight: value
+});
+
+/**
+ * Use to reduce the amount of code
+ * @param {object} color Object with keys that defining colors using the Red-green-blue (RGB) model
+ */
+const setColorTamplate = color => `rgb(${color.r}, ${color.g}, ${color.b})`;
```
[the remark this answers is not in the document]
TreTuna: please see my other comment about the two functions named `setColorTamplate`. I think we should rework this into a shared function.

src/renderer/components/ColorPalette/ColorPalette.js
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * This is Hook that lets add React state "indexFocusButton" to functional components
+   * @param {object} [initialState=selected] - Sets initial state for "indexFocusButton"
+   */
+  const [indexFocusButton, setIndexFocusButton] = useState(selected);
+
+  /**
+   * Use variable to reduce the amount of code
+   */
+  const colorForFocusButton = {
```
TreTuna: Why are we creating this variable at all vs just using `...palette[selected]` where you are using `...colorForFocusButton`?

src/renderer/screens/ColorPalette/PickerColorButton.js
0957758592: @TreTuna  

In the last `refactor` [commit](https://github.com/keyboardio/Chrysalis/commit/ecedbf5cc77129bd6b8b66575fb7379eecff4e41) there were makes changes in [`ColorPickerButton`](https://github.com/keyboardio/Chrysalis/blob/ecedbf5cc77129bd6b8b66575fb7379eecff4e41/src/renderer/components/ColorPalette/PickerColorButton.js) from custom button as `div` to [Floating Button ](https://material-ui.com/components/buttons/) from Material UI. 
And the standart width in Floating Button is 56px was changed as 

```
  fab: {
    width: 60,
    height: 60
```

src/renderer/screens/ColorPalette/index.js
0957758592: @TreTuna 

Sorruy, Its my fault
In `ColorPalette.js` there were makes changes to dinamic height depends of the width of screenSize - it will makes height of colorPalette more flexible, but didn't set `minHeight` yet. - my fault. Will update for today.

src/renderer/components/ColorPalette/PickerColorButton.js
```diff
@@ -0,0 +1,124 @@
...
+   * Change "anchorEl" in functional component state to close Color Picker
+   */
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  /// Set the value to open (close) Popover element
+  const open = Boolean(anchorEl);
+  /// Set the value to id Popover element
+  const id = open ? "simple-popover" : null;
```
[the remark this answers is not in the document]
0957758592: This were used as defaullt API from [`react-color` ](https://casesandberg.github.io/react-color/)modules.
So now using simple `id={"simple-popover"}`

src/renderer/screens/Editor/Editor.js
```diff
@@ -604,7 +603,7 @@ class Editor extends React.Component {
               )}
             </ToggleButtonGroup>
             <div className={classes.grow} />
-            <FormControl className={classes.layerSelect}>
+            <FormControl>
```
TreTuna: Please add this class back

TreTuna:
I'm not sure how I feel about the double borders in this area. Maybe we remove the border around the palette buttons? The double border on the left feels... icky.

image

obra:
Yeah. Removing the interior border sounds about right.

TreTuna:
This is looking good @0957758592. I have to go through one more time tonight, but feeling pretty good about it right now I think. @algernon how are you feeling here?

PR #396, opened 2019-05-30
(#385) enhancement: main window changes for usability 

0957758592:
Signed-off-by: Roman Danilov

- Added new Cancel Button to dismiss changes.
- Moved layers menu to the right side,
- Moved HomePage link into main (hamburger) Menu
- Show bottom menu in first run application (hide bottom menu clicking the key)

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -222,7 +222,6 @@ class App extends React.Component {
               connected={connected}
               pages={pages}
               device={device}
-              cancelContext={this.cancelContext}
             />
             <main className={classes.content}>
               <Router>
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -222,6 +222,7 @@ class App extends React.Component {
               connected={connected}
               pages={pages}
               device={device}
+              cancelContext={this.cancelContext}
             />
             <main className={classes.content}>
               <Router>
--- a/src/renderer/components/CancelChangesButton.js
+++ b/src/renderer/components/CancelChangesButton.js
@@ -1,0 +1,63 @@
+// -*- mode: js-jsx -*-
+/* Chrysalis -- Kaleidoscope Command Center
+ * Copyright (C) 2018, 2019  Keyboardio, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import React from "react";
+import PropTypes from "prop-types";
+import { withStyles } from "@material-ui/core/styles";
+import Tooltip from "@material-ui/core/Tooltip";
+import Fab from "@material-ui/core/Fab";
+import CloseIcon from "@material-ui/icons/Close";
+
+CancelChangesButton.propTypes = {
+  classes: PropTypes.object.isRequired,
+  cancelContext: PropTypes.func.isRequired
+};
+
+const styles = theme => ({
+  fab: {
+    position: "fixed",
+    justifyContent: "center",
+    bottom: 0,
+    right: theme.spacing.unit * 9,
+    margin: theme.spacing.unit
+  }
+});
+
+/**
+ * Reactjs functional component that create button for cancel changes
+ * @param {object} classes Property that sets up CSS classes that adding to HTML elements
+ * @param {function} cancelContext Function that cancel context. Parameter - "true"
+ */
+function CancelChangesButton(props) {
+  const { classes, cancelContext } = props;
+
+  return (
+    <Tooltip placement="top" title={props.children}>
+      <Fab
+        color="secondary"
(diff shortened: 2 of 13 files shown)


TreTuna:
Just to reference here, the discussion around these changes took place here: https://github.com/keyboardio/Chrysalis/issues/385

I think we have a break down of communication here as it seems there is work being done that was never finished in the discussions. From what I see, the last portion of the discussion was to *not* implement a couple of these changes. I'm not going through all the code here right now and making individual comments as there needs to be more discussion around these changes before they go any farther. Here are my notes from the working around current UX of all of this:

> Added new Cancel Button to dismiss changes.
The discussion around this did not come to a conclusion and while I think this button looks good, the UX around this is wrong. As @algernon brought up in his comment, having two FABs can be done in Material-UI, but recommended against. This is a case where we can do it, but the current implementation needs a little bit of work.

The problem here is that now a user can make a change to my layout and then navigate away without confirming or canceling my changes. Users can now go change a button, navigate to the firmware screen (still with a change cued) and flash firmware or goto keyboard select and disconnect from the board and while I don't have my other keyboard here at home, I suspect they may be able to even connect to an entirely different keyboard while these changes are still cued to the first. 

The simplest solution I can think of real quick to this would be to make them both have the same action, canceling the newest changes. While having two buttons that do the same functionality feels weird, the menu button doesn't feel **too** bad on this as you have to either cancel or save your changes before navigating away. This should be discussed more and decided on before the work is done though.

> Moved layers menu to the right side,
The current master branch already has this change included. It looks like you made here was to remove the margin-right setting that was there, which honestly feels too close to the other buttons. I would be ok dropping the number there by half, but not removing it completely. 

> Moved HomePage link into main (hamburger) Menu
This was also discussed and the last words about it were that we were not wanting to make this change. 

Also, this implementation only gives me a homepage link and not the other links that I had formerly in the Keyboard menu. Not only that, when I'm connected to my Model01 the Homepage link takes me to the homepage of the Dygma Raise (looks like it's hardcoded in like this) and gone are all three links to the Keyboardio sites. My personal preference would be to fully revert this change.

> Show bottom menu in first run application (hide bottom menu clicking the key)
This is spot on and looks great! Looks like you implemented it exactly how Algernon had talked about it.

0957758592:
> Just to reference here, the discussion around these changes took place here: #385
> 
> I think we have a break down of communication here as it seems there is work being done that was never finished in the discussions. From what I see, the last portion of the discussion was to _not_ implement a couple of these changes. I'm not going through all the code here right now and making individual comments as there needs to be more discussion around these changes before they go any farther. Here are my notes from the working around current UX of all of this:
> 
> > Added new Cancel Button to dismiss changes.
> > The discussion around this did not come to a conclusion and while I think this button looks good, the UX around this is wrong. As @algernon brought up in his comment, having two FABs can be done in Material-UI, but recommended against. This is a case where we can do it, but the current implementation needs a little bit of work.
> 
> The problem here is that now a user can make a change to my layout and then navigate away without confirming or canceling my changes. Users can now go change a button, navigate to the firmware screen (still with a change cued) and flash firmware or goto keyboard select and disconnect from the board and while I don't have my other keyboard here at home, I suspect they may be able to even connect to an entirely different keyboard while these changes are still cued to the first.
> 
> The simplest solution I can think of real quick to this would be to make them both have the same action, canceling the newest changes. While having two buttons that do the same functionality feels weird, the menu button doesn't feel **too** bad on this as you have to either cancel or save your changes before navigating away. This should be discussed more and decided on before the work is done though.
> 
> > Moved layers menu to the right side,
> 
> The current master branch already has this change included. It looks like you made here was to remove the margin-right setting that was there, which honestly feels too close to the other buttons. I would be ok dropping the number there by half, but not removing it completely.
> 
> > Moved HomePage link into main (hamburger) Menu
> 
> This was also discussed and the last words about it were that we were not wanting to make this change.
> 
> Also, this implementation only gives me a homepage link and not the other links that I had formerly in the Keyboard menu. Not only that, when I'm connected to my Model01 the Homepage link takes me to the homepage of the Dygma Raise (looks like it's hardcoded in like this) and gone are all three links to the Keyboardio sites. My personal preference would be to fully revert this change.
> 
> > Show bottom menu in first run application (hide bottom menu clicking the key)
> 
> This is spot on and looks great! Looks like you implemented it exactly how Algernon had talked about it.

### here is a view of functionality(partly) issue #385:
- started app without hidden bottom panels;
- moved `HomePage` link to the `hamburger menu;
- displayed red button for changes cancellation (next to the `save changes button`)

https://www.loom.com/share/48d97489128c413d9508210134d504b9

0957758592:
Updated:
1) Pulled from master
2) Added: never hide the bottom menu. Automatically select a key at launch and have this shown in the bottom menu.

TreTuna:
One of the things that's going on here is a lack of communication. I see now that you reverted the entire thing to master, which nixed most of my comments. I thank you for that. Let's get things resolved and pushed through before adding new changes, please. If the addition of "never closing" the bottom menu hadn't been introduced here we may have had a path, but it seems each time there's something added that was explicitly stated we didn't want to do.

0957758592:
@TreTuna 
Thank you for the comments.
Im appreciate you for helping and letting me deeply understand entire project, users expectations and functionalities.

I'm really a little bit confused now with the steps should be done with this issue.

It seemed that everything was clear to me.

There were would be better to reproduce the steps should be done e.g. 

1) Add function...
2) Remove implementation..
3) Refactor..
etc..

Will be discussed with @mattvenn for today to make me clear )

0957758592:
Closed for the [reason](https://github.com/keyboardio/Chrysalis/issues/385#issuecomment-504371780)

PR #398, opened 2019-06-11
(#382) multiple select LEDs

0957758592:
### **Update step for #381** 

**Enhancements:**
1) bottom menu doesn't hidden after starting app, and after switching between `colored` and `keys` panels (depends to #396)
2) `button-Pallete` disabled in case `color-button` not on active stage

- to reproduce Multiple selection is just should activate (make a click) a `color-button` and than click to any keys you would like to change a color (except current key => cause it will hide the panel)
- to  turn off `color-button` - just click once again on it
- to know what the color is on current key - press `CTRL` and click the key - `color-button` turning on (as box-shadows around)

Also added a prevention for `color-button` - in case using `CTRL` and click to the `color-button` nothing will changes (its lock to prevent during clicking the key)

********************************
**Will be updated:**

- Click any the key - to allocate it.
- To hide a bottom panel - click to the same key again.

This case avoid to make a current (allocated) key become a coloured 

*To resolve it and hide bottom menu I have suggest a next functionality: 
just click outside the keyboard, 
and to turn bottom panel back - ckick to any key again (the function as is).

### DIFF
--- a/src/renderer/components/ColorPalette/ColorButton.js
+++ b/src/renderer/components/ColorPalette/ColorButton.js
@@ -21,6 +21,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { withStyles } from "@material-ui/core/styles";
 import Button from "@material-ui/core/Button";
+import ColorizeIcon from "@material-ui/icons/Colorize";
 import { setButtonSizeTamplate } from "../../../renderer/utils/setTemplates";
 
 ColorButton.propTypes = {
@@ -29,7 +30,8 @@ ColorButton.propTypes = {
   setIsFocus: PropTypes.func.isRequired,
   index: PropTypes.number.isRequired,
   color: PropTypes.object.isRequired,
-  disabled: PropTypes.bool.isRequired
+  disabled: PropTypes.bool.isRequired,
+  isSelected: PropTypes.bool
 };
 
 const styles = theme => ({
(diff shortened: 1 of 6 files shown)


TreTuna:
Thanks a lot, I really like the UX of painting side of this so far! There would be one thing I believe is a must-have for clarity to users: When a user has a key selected already, and then you click on a new color, it should change the color of the currently selected key as well as allowing you to "paint" other keys like it currently does. 

Right now it's confusing that when I select a key and then select a new color it doesn't change that key. My mind first goes to "Wait, what? How do I change the key I have selected then?" Before I realize I have to either a) color a different key first then come back to the selected one. b) Select a different key, color, then click on the one I wanted to change in the first place.

0957758592:
> Thanks a lot, I really like the UX of painting side of this so far! There would be one thing I believe is a must-have for clarity to users: When a user has a key selected already, and then you click on a new color, it should change the color of the currently selected key as well as allowing you to "paint" other keys like it currently does.
> 
> Right now it's confusing that when I select a key and then select a new color it doesn't change that key. My mind first goes to "Wait, what? How do I change the key I have selected then?" Before I realize I have to either a) color a different key first then come back to the selected one. b) Select a different key, color, then click on the one I wanted to change in the first place.


**Here is a view of last update implementation:**
- resolve collapsing with second clicking the same key to hide the `bottom-panel` and setting color
- added: `bottom-panel` clicking outside the `key panel`
- set the color to selected key (at first time)

https://www.loom.com/share/5f7a448783cc4762ab5f98568d78aa05

pull request was updated [here](https://github.com/keyboardio/Chrysalis/pull/398/commits/24add6f2078b9c21b566273d40693f6dc05276f8)

TreTuna:
Alright, the functionality is looking pretty good. I still need to go through the code more, however, there's something else that needs to be taken care of before we can move forward here anyway.

> bottom menu doesn't hidden after starting app, and after switching between colored and keys panels (depends to #396)

If this PR is dependent upon that one, we need to get that finished first. There are several things that I commented on that were _not_ supposed to be changed, but were. I've left many more comments on individual lines for these changes there.

0957758592:
**Updated:**
- removed all additional implementation
***********

* to reproduce Multiple selection is just should:
1) Click on any key you would like to change a color to activate bottom `color-menue`, 
2) Click to a `color-button` to activate it, 
3) Change the color with `color-Picker`,

During `color-button` on active state - make a multiple selection another key just clicking to them
* to  turn off `color-button` - just click once again on it
* to know what the color is on current key - press `CTRL` and click the key - `color-button` turning on (as box-shadows around)

> Also added a prevention for `color-button` - in case using `CTRL` and click to the `color-button` nothing will changes (its lock to prevent during clicking the key)

Issue #372, opened 2019-04-12
Show proper characters in the select boxes when shift is checked


TreTuna:
**Is your feature request related to a problem? Please describe.**
Currently, when you set a key to either a number or punctuation keycode, then click on the checkbox to hold shift, the key itself shows the proper character, but the key select box still shows the non-shifted characters. ie: With shift-clicked, the char select shows a semicolon, though the key itself shows a colon, which is the character that is output by shift+semicolon.

**Describe the solution you'd like**
The solution I can think of at the moment would be to change the characters in the key select boxes to show the proper character depending on if the shift checkbox is checked or now. ie: if the shift checkbox is marked true, the key selector for a semicolon would turn into a colon.

**Describe alternatives you've considered**
none at this time

**Additional context**

<img width="222" alt="Screenshot 2019-04-11 22 41 09" src="https://user-images.githubusercontent.com/16496746/56014669-13a11e80-5cab-11e9-8f76-f6c3dc3986e4.png">

<img width="222" alt="Screenshot 2019-04-11 22 41 32" src="https://user-images.githubusercontent.com/16496746/56014674-18fe6900-5cab-11e9-8ee1-719a95a392de.png">

obra:
This makes sense, though we'll need to figure out something clever when
doing internationalization.
ᐧ

algernon:
We have redesigned the keymap editor since this issue was reported, and the problem does not exist in the new version:

![Screenshot from 2022-05-17 12-36-07](https://user-images.githubusercontent.com/17243/168792485-d9d57c2f-1300-49c5-85ae-4faf589bc7c1.png)

PR #421, opened 2019-08-11
#372: fixes Show proper characters in the select boxes when shift is checked

0957758592:
fixes: #372 

Signed-off-by: Roman Danilov


0957758592:
https://www.loom.com/share/b50543c73b4c41b9afa2f2a2ed99a213

obra:
This actually feels pretty weird to me. I'd expect that the text of -all- the buttons would change when I selected the modifier keys, rather than having the button magically change when I selected it.

PR #431, opened 2019-10-10
Formatting for Homebrew commands

brow:
The existing markdown renders as a single line, which errors when pasted into a terminal:

```
$ brew update && brew upgrade brew cask install chrysalis
Already up-to-date.
Error: No available formula with the name "brew"
```


obra:
Thanks!

PR #494, opened 2020-04-02
Add Num Lock key

replicaJunction:
Adds a label and keycode for the Num Lock key.

I've validated the keycode using the "Unknown keycodes" entry in Chrysalis and confirmed it sends the correct Num Lock code on Windows 10.

### DIFF
--- a/src/api/keymap/db/miscellaneous.js
+++ b/src/api/keymap/db/miscellaneous.js
@@ -25,6 +25,13 @@ const MiscellaneousTable = {
       }
     },
     {
+      code: 83,
+      labels: {
+        primary: "NumLK",
+        verbose: "Num Lock"
+      }
+    },
+    {
       code: 71,
       labels: {
         primary: "ScrlLK",


algernon:
This looks good to me. The only thing I'm torn about is the group where `Num Lock` should go. It makes sense for the `miscellaneous` group, because we have `Scroll Lock` there. But we have `Caps Lock` in `punctuation`, and on a traditional keyboard, `Num Lock` is typically on the numpad, so the `numpad` group would also fit.

@obra what's your opinion about this? Where should the `Num Lock` key go? `Miscellaneous`, `Numpad` or some other group?

obra:
I'd probably  stick all three of them in the same group, along with Shift,
Control, Alt, GUI, tc.

algernon:
> I'd probably  stick all three of them in the same group, along with Shift, Control, Alt, GUI, tc.

So all of them in `Modifiers`? If so, I'll merge this as-is, and move them all there in a followup.

obra:
Maybe rename modifiers to "Modifiers and Locks"?

that's not perfect, but an improvement i think

PR #505, opened 2020-04-26
Debugging configuration, update readme, add sample debugging configurations.

pm0u:
This PR adds sample debugging configurations under a `dev/sample` folder (would this be the proper/desired location?) and and ENV file that sets proper environment variables for the main process, and allows a remote debugging port to be passed to the renderer process.

Few gotchas:
1. The launch for main provided doesn't _always_ seem to connect in time to catch breakpoints in `src/main/index.js`. Sometimes it will, other times it seems to attach after this code has been executed.
2. It is possible to use `--inspect-brk=$PORT` under ELECTRON_ARGS to break on the first line (and then ideally you could connect debugger before any code in `src/main/index.js` is executed)... _but_ this seems to break on the first line of some internal electron process, rather than the app.

_A workaround for these could be to throw a `debugger;` line in `src/main/index.js` if you need to debug that file, and then hopefully execution would sit until the debugger connects, unsure on that one._

~~3. The renderer process doesn't seem to map breakpoints within class components properly. (more reason to use functional components? :smile: ) Any breakpoint set within the component is mapped to the class declaration line, and then following execution leads to a rabbit hole of inner React modules. I am not sure how to get past this, and I couldn't find anything about this in searching.~~

Re: # 3 -- This seems to be related to HMR, if a refresh is forced once the app has opened (Ctrl + r), then breakpoints are mapped properly. Unsure why.

In VSCode it is also possible to launch with debug params in an external console (supposedly) but when i tried this, it just launched my terminal and never acutally launched the app within that terminal window. I think this would be ideal for the main process, rather than launching within VSCode's internal terminal (which can be hard to use).

## Reference

[electron-webpack env vars]()
[debugging issue 1](https://github.com/electron-userland/electron-webpack/issues/76)
[debugging issue 2](https://github.com/electron-userland/electron-webpack/issues/80)


dev/sample/.vscode/launch.json
```diff
@@ -0,0 +1,58 @@
...
+    // Configuration adapted from https://github.com/electron-userland/electron-webpack/issues/80.
+    {
+      "name": "Electron Webpack: Launch Main",
+      "type": "node",
+      "request": "launch",
+      "cwd": "${workspaceFolder}",
+      "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron-webpack",
+      "args": ["dev"],
+      "env": {
+        //"DEBUG":"*",
```
pm0u, line 35: I left this line commented as it _may_ be useful. This outputs extremely verbose output from electron/electron-webpack.

dev/sample/.vscode/launch.json
```diff
@@ -0,0 +1,58 @@
...
+        "ELECTRON_DISABLE_SECURITY_WARNINGS": "1"
+      },
+      "autoAttachChildProcesses": true,
+      "console": "integratedTerminal",
+      "outputCapture": "std",
+      "sourceMaps": true,
+      "smartStep": true,
+      "showAsyncStacks": true,
+      "timeout": 30000
```
pm0u, line 44: This timeout will vary, but it needs to be _longer_ than the amount of time it takes for the electron app to launch so that VSCode does not give up. Perhaps i will note this..

pm0u: noted.

obra:
@algernon - go ahead and merge at your leisure, then :)

PR #515, opened 2020-05-15
Add "Right Alt" in parens to AltGr

markis:
Currently the "Modifiers & Locks" keymap has "LEFT ALT" and "ALTGR"

image

This PR updates the UI to add Right Alt in parens next to AltGr, for those who are unsure of what is AltGr.

### DIFF
--- a/src/api/keymap/db/modifiers.js
+++ b/src/api/keymap/db/modifiers.js
@@ -66,7 +66,7 @@ const ModifiersTable = {
       code: 230,
       labels: {
         primary: "RAlt",
-        verbose: "AltGr"
+        verbose: "AltGr (Right Alt)"
       }
     },
     {


obra:
Both are technically correct. AltGr is the standard European spelling, as in Europe it’s frequently used for typing alternate or accented glyphs.

markis:
Interesting, how would you feel about adding a "Right Alt" button that's mapped exactly the same as "AltGr"? That way it appeases both audiences.

obra:
Maybe putting (Right Alt) in parents in the key chooser entry for AltGr?

markis:
That's a great idea! I updated the code, title and description to reflect the change.

obra:
Looks good to me! @algernon - Does this work for you?

algernon:
This completely fell off my radar, sorry!

Yep, looks all good, merging. Thanks!

PR #516, opened 2020-05-17
Clearing a layer causes Chrysalis to crash if there is no color map

pm0u:
## To reproduce

1. connect a keyboard with no colormap / no colormap set (atreus2 worked for me)
2. In the editor screen, attempt to clear any layer.

## Screenshot

crash_clear_layer

### DIFF
--- a/src/renderer/screens/Editor/Editor.js
+++ b/src/renderer/screens/Editor/Editor.js
@@ -476,7 +476,7 @@ class Editor extends React.Component {
           .map(() => ({ keyCode: 0xffff }));
       }
 
-      let newColormap;
+      let newColormap = [];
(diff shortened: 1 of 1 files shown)


pm0u:
Unsure if a similar fix should be applied to keyMap? not sure I understand fully whether it would be possible to not have a keyMap.. but if it is, the same issue would occur as it would be set to undefined when anywhere else is expecting only the possiblity of an array

obra:
Bulletproofing against missing data seems to be a good thing generally,
even if the possibility is unlikely.

obra:
I'm happy to have that other fix as a separate PR. I'm merging this one.

Thanks!
jesse

PR #553, opened 2020-08-13
Change Model01 URL to Kickstarter description

Rom1deTroyes:
Keyboard.io Model01 is no longer availlable on the shop, so the link seems not acurate.

Signed-off-by: Rom1deTroyes Rom1deTroyes@users.noreply.github.com


obra:
Thanks for the PR. For the Model 01, we still want to be redirecting users to a site we control, rather than Kickstarter, which we can't even edit.

PR #632, opened 2021-01-17
Issue 609 nonexistent layers

MagicKriss:
Fix for bug https://github.com/keyboardio/Chrysalis/issues/609 - "Chrysalis + Atreus +  Move to layer exposes nonexistent layers"

Husky's pre-commit hooks would nag me about another file I didn't touch, so I ran prettier.

### DIFF
--- a/src/renderer/components/ColorPalette/ColorButton.js
+++ b/src/renderer/components/ColorPalette/ColorButton.js
@@ -69,7 +69,8 @@ function ColorButton(props) {
   ///Checks background is white or not
   const isWhiteColor =
     color.r >= minWhiteColorValue &&
-    (color.g >= minWhiteColorValue && color.b >= minWhiteColorValue);
+    color.g >= minWhiteColorValue &&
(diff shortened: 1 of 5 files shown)


obra:
Thanks!

@algernon can you please review tomorow?

algernon (changes requested):
This looks good for the most part (thank you!), there's one thing that I think isn't quite ok: with this change, with removing the `min`/`max` properties from the target layer `<Input>` element, the user would be able to - seemingly - set a layer that doesn't exist, because the index gets clamped down to fit into the interval in `onTargetLayerChange`.

Instead, we should still have the `min`/`max` on the input, but use `getMaxLayer` to set `max`. Other than this, approved.

Since this is a fairly annoying problem, I can merge this as-is, and patch it up post-merge.

MagicKriss:
@algernon I have added back max/min. Please take a look

algernon (approved):
Perfect, thank you!

PR #670, opened 2021-03-03
[Fix regression] Larger  key legends for single characters

### LINKED ISSUE #669: [Atreus] Key labels so small that they cannot be read
**Describe the bug**
I am a 30 year old with no visual impairment, however since the latest update, the font size of some of the keyboard legends became so small that I can no longer read it. 

**To Reproduce**
1. Connect a keyboard
2. Open Chrysalis
3. Press connect

**Expected behavior**
The font size of the key labels is large enough to be read on stock Chrysalis on a stock retina MacBook.

**Screenshots**
# Before (Chrysalis v 0.7.9)
image

# After (Chrysalis v 0.8.3)
image

 - OS: MacOS Catalina
 - Chrysalis Version: 0.8.3

**Additional context**
Please consider you might have a different screen DPI, and what is legible in your screen might not be possible to read in other people's displays. Also, all other text in the application is of a legible size, it is only some key labels that are too small to read now. It did not used to be this way and hence why it seems to me like a regression.

Here is a photo of my thumb on the screen for a sense of scale image

**Codewise**
I inspected with dev-tools and codewise, the bug is that key legends now fail to be assigned the `shortLegend` className, which they used to before, and which makes their fontsize `16px` rather than `8px` as it is now.

image

The culprit seems to be that the following piece of code in `src/api/hardware-keyboardio-atreus2/components/Keymap.js` has gone missing 
image

RodrigoRoaRodriguez:
Fixes #669.

There was a regression when the new UI code was implemented and a piece of
code that applied conditional formatting went missing.

# Before
image

# After

# How it looked on version 0.2.9
image

Signed-off-by: RodrigoRoaRodriguez <rodrigoroarod@gmail.com>

### DIFF
--- a/src/api/hardware-keyboardio-atreus2/components/Keymap.js
+++ b/src/api/hardware-keyboardio-atreus2/components/Keymap.js
@@ -101,7 +101,9 @@ class Keymap extends React.Component {
       let textColor = "#ffffff";
       const buttonColor = "transparent";
 
-      let legendClass;
+      let legendClass = "";
+      if (key && (db.format(key).main || "").length <= 1)
+        legendClass = "shortLegend";
       if (key && key.code == 0) textColor = "#888888";
 
       return (


RodrigoRoaRodriguez:
Hello, could you please help me with the DCO? I used my primary e-mail address but the test expects me to use my work e-mail instead (which I do not want to use, since this PR is a free time activity for me).
image

Rom1deTroyes:
@RodrigoRoaRodriguez the email needed must match the one from your Github account, and the name part too...
... I had the same trouble, and now use my Github email for committing on Github project.

Maybe adding this line on the commit text can pass the DCO test :
```
Signed-off-by:  Rodrigo Roa Rodríguez <RodrigoRoaRodriguez@users.noreply.github.com>
```
(Or you can change your email privacy just for the test, and back to normal after ?)
image

RodrigoRoaRodriguez:
The thing is: _it already matches the one from the github account_. That is why I posted the image above. I do not use my work email for Github, nor github at my work, so the DCO is probably nabbing it from my global git configuration rather than my Github account. 

I will try to figure out a way to trick the DCO into using the e-mail address associated with my GitHub account. Any suggestions on how to do this are welcome.

obra:
I can also manually force the DCO based on your statement. Thanks so much for the patch!

PR #766, opened 2022-03-30
Initial changes to move us to functional components so we can use react hooks

obra:


src/renderer/screens/Editor/components/OnlyCustomScreen.js
```diff
@@ -25,77 +25,58 @@ import CardContent from "@mui/material/CardContent";
...
-            >
-              {i18n.t("editor.onlyCustom.fixItButton")}
+  return (
+    <div sx={{ display: "flex", justifyContent: "center" }}>
+      <Portal container={props.titleElement}>
+        {i18n.t("app.actionRequired")}
+      </Portal>
+      <Card
+        sx={{
+          margin: 4,
```
algernon, line 56: This should probably be `margin: (theme) => theme.spacing(4)`, rather than `margin: 4`.

obra:
that’s what the 4 means. sx does things in theme spacing units automatically.

algernon:
There's a lot I like about this change, moving to functional components is one of them. Moving away from the deprecated styling is also a welcome step. I'm not sure I'm sold on doing both at the same time. I'm not going to try and persuade you to split this PR to move the styling stuff to a separate one, but it would make reviews slightly easier if we did the two separately.

I'll resolve the conflict, and merge this, because they're all great improvements. Thanks!

algernon:
> that’s what the 4 means. sx does things in theme spacing units automatically.

Oh, that's nice. Thanks for the clarification!

PR #770, opened 2022-04-01
Style and functional component refactoring

obra:


obra:
@algernon I think this is now ready for review. It's a whole bunch of stuff but should be pretty straightforward

PR #773, opened 2022-04-10
Refactor the KeyboardSelect screen

obra:
* break out more components into their own files
* remove old style code
* this required touching the atreus device image a bit, too


algernon:
The changes to the Atreus image have an unwanted side-effect, it no longer displays full-size while editing. See below for comparisons:

## Editing

### Current master

Screenshot from 2022-04-19 17-26-22

### Refactored

Screenshot from 2022-04-19 17-27-55


## Keyboard select

### Current master

Screenshot from 2022-04-19 17-26-49

### Refactored

Screenshot from 2022-04-19 17-27-20

## Notes

I do like the change on the keyboard select screen. On the Editor, though, the keyboard image is too small, and the fonts are wrong, too (Roboto after the refactor, Source Code Pro on master; size and weight are wrong too: 16 & 400 vs 8 & 700).

As this is a reasonably localized change, I'd rather not merge it just yet, but figure out how to fix it first.

algernon (changes requested):
I suppose I should've posted my previous comment in a review... see [here](https://github.com/keyboardio/Chrysalis/pull/773#issuecomment-1102797482) for more information about why I'd like to request changes.

obra:
I actually really dislike Source Code Pro for keyboard labels. The size
changes were intentional, but it looks like the multi-character legend size
looks like it ended up too big.

(Everything else you flagged does look like a bug. Thanks for the catch.

obra:
@algernon - this is once again ready for review.

algernon:
I can be convinced about a font change, not particularly passionate about SCP. Taking another look now, thanks!

algernon (approved):

algernon:
Looks much better now!

There's one _tiny_ little thing left: the keyboard select screen used to hide labels on the keyboard, and it seems to show them now (at least for the Atreus). I think we should be showing an empty keyboard there, the labels aren't readably anyway, and they are a bit... distracting?

Screenshot from 2022-04-19 21-14-27

I'm happy to merge this PR in its current state, and fix the preview later, mind you. Marked it approved, feel free to either fix the select screen preview, or merge, or let me know which one you prefer.

PR #774, opened 2022-04-10
Restyle Prefs, extracting components to their own files

obra:


obra:
The reason I restored the dropdowns to full width is *primarily* that the
previous styling requires extensive fighting against the toolkit. It's not
just the Material UI guidelines but the actual MUI framework. It really,
really, really does not want you to have "small" dropdowns with side
labels.

The previous layout also has a serious usability issue in that labels are
incredibly distant from the widgets they control. With left-side labels,
that was the only way to get things lined up properly, but it seriously
hurt usability.

It might be possible to make the dropdowns not fill the entire column, but
that looks even weirder with tons and tons of whitespace and a narrow
column of content.

I'm happy to fix up the margins, though.

algernon:
Mhm, I understand your points, and I do not disagree. I still dislike how things look after this PR, but as long as we agree to improve on that in the not too distant future and not leave it as-is, I'm not opposed to merging the PR.

obra:
@algernon this is once again ready for review

algernon (approved):

algernon:
OOOOH. I'm liking this direction very much. Still room to improve here and there, but it is more than good enough to calm my worries.

Merging, thanks!

PR #885, opened 2022-06-01
Update dfu-util to 0.11

obra:
For linux and windows, use Arduino's builds.

For macos, use new native builds for each architecture, partially because it's right and partially to work around electron-builder issues.

After this is tested and merged, we can switch back to the 'regular' build of electron/universal (v1.2.1) and then come back up to electron 18.


algernon (changes requested):
Unfortunately, the `dfu-util` linux builds were built against a glibc that's too new. Our current one requires at least glibc 2.17, which is ancient, but it pretty much guarantees our dfu-utils will run on any distro.

With this PR, we're upping that to glibc 2.34, released late last year, and it has not made it back into distributions like Debian Stable, which has 2.31. Ubuntu Bionic LTS, which is supported until April next year, has glibc 2.27. I have no idea what LTS distros ship with, but I imagine they'll be using a glibc _much_ older than 2.34.

We'll need to figure out something about this, because I can't merge this as-is, because that breaks flashing on Linux on anything but pretty much bleeding edge distros.

obra:
ack. ok.

algernon:
FWIW, a useful commandline magic I used to check what glibc dfu-util needs:

```
❯ objdump -trw static/dfu-util/linux-x64/dfu-util | sed -ne 's,^.*@\(GLIBC_.*\)$,\1,p' | sort -ruV | head -n1
GLIBC_2.34
```

gergelypolonkai:
Although it would make the `dfu-util` binary pretty big, wouldn’t it work if you add a statically linked version? That should work on pretty much any distro.

algernon:
Static linking is certainly an option, and probably the right one, too. It's just additional burden on us, having to build it, and automate that, and so on. If we build it, we likely want to static link it against something other than glibc (musl comes to mind; it builds against musl, alpine has it packaged without patches).

obra:
Rebuilt and repressed the linux build.

algernon (approved):
Confirmed working on Linux (Debian stable) & macOS on Apple M1.

PR #961, opened 2022-06-24
Speed up layout selector and lay groundwork for not needing main process cldr code.

obra:


obra:
@algernon it may make sense to cherry pick the first commit from this. Up to you. 

I'm not *quite* sure how to properly use these new json files from the renderer side. Do you have ideas or advice? (I'm happy to hand off to you)

algernon:
> @algernon it may make sense to cherry pick the first commit from this. Up to you.

I'll cherry pick the first commit indeed, that looks solid as it is, while the second will need some thinking, imo.

> I'm not _quite_ sure how to properly use these new json files from the renderer side. Do you have ideas or advice? (I'm happy to hand off to you)

I'm not sure. I don't think separate files for each locale is going to work, unless we package those files up (in which case they shouldn't be in `src`), and lazy-load them at runtime. Lazy loading would make sense, there's no reason to keep all that data in memory, or even load it all at the same time. We can also filter it down to languages we support (en and nl at the moment), which makes the size go down from ~9.3mb to 32kb, from 803 files to 2.

When we're down to 2 files, ~16k each, we can just import them at the few places we need to - which involves moving them out of `src/api`, and into `renderer`.

I'm happy to take the 2nd part further.

algernon:
I just had an idea: we can leave the generated json files where they are, and we don't necessarily need to filter what we generate, either. We'll import the two we need, and the rest simply won't be included. If they're not imported, they don't end up in the webpack either, so no space wasted, except a little during build.

algernon:
Verified that unused mappings won't end up in the webpack, so we are free to generate all of them. Also updated the PR a little, fixing a bug (supported layouts weren't memoized per language, so changing the language of Chrysalis didn't change the translations of the supported layouts until Chrysalis was restarted), and it now makes use of the pre-generated language maps, rather than calling into the main process to do a lookup with CLDR.

With these changes, it's good to go in, I believe, so I'll do just that, and merge.

Thanks for the groundwork @obra!

obra:
That seems reasonable. If we *can* load them dynamically, so much the better. It means less we need to change when we add a locale

PR #1045, opened 2022-09-12
Refactor and reorganize mainmenu, starting to split it into smaller bits

obra:


algernon (changes requested):
I'm not sure I'm happy about where the `Preferences` menu went:

Screenshot from 2022-09-12 21-26-57

I feel it's too close to `Exit`, and it feels a bit... weird to be stuck at the end of a bunch of (mostly external) links. Perhaps a divider could work? No need for a title for the separate section, just a divider before Preferences? I might feel better about having it next to Exit then, too.

The PR also broke the logic that pushed us to the layout editor upon connect, and connecting to a Model100 I land on the "Welcome to Chrysalis" screen:

Screenshot from 2022-09-12 21-29-55

As you can see, the keyboard *does* respond.

Interestingly, landing on the Welcome screen, the `Layout Cards` menu is still available (this might be the case on master too, I have not verified):

Screenshot from 2022-09-12 21-30-33

obra:
Thanks. will rework

obra:
Closing for now. Will reimplement later

PR #1102, opened 2022-09-27
Change how we let users perform a full factory reset, making the option more prominent.

### LINKED ISSUE #1085: Let me easily return to the default keyboard
Yes.  When I tried to change a key, I lost all the other keys (they now do nothing—only the key I changed works).  

A clear and simple way to restore my keyboard to its default keys.

Throwing my Keyboardio out the window.

**Additional context**
I expected a keyboard as good as the Keyboardio would come with software that was equally as good.  It didn't.  Chrysalis is *terrible*.  If I gave you a list of the frustration in my 30 minutes trying to do something simple with this software I'd be here all morning.  Now I'm left with a keyboard that doesn't work.

I'm not even sure Chrysalis is worth fixing.  If I were you, I'd get a new software developer and have them start from scratch.

obra:
image

Fixes #1085 

@algernon - please test this one out before merging. Just in case

### DIFF
--- a/src/renderer/i18n/en.json
+++ b/src/renderer/i18n/en.json
@@ -406,6 +406,9 @@
     },
     "firmwareSources": "Source code",
     "calloutTitle": "Important!",
+    "factoryResetTitle": "Factory reset",
+    "factoryResetDescription": "Erase all settings layouts on your keyboard, restoring it to factory defaults.",
+"factoryResetWarning": "Selecting this option will not backup your current layout or settings.",
     "dialog": {
       "selectFirmware": "Select a firmware",
       "firmwareFiles": "Firmware files",
@@ -415,7 +418,6 @@
       "error": "Error flashing the firmware",
       "success": "Firmware flashed successfully!",
       "button": "Update",
-      "factoryResetButton": "Factory reset",
       "notifications": {
         "title": "Problem encountered while flashing",
         "enter": {
--- a/src/renderer/i18n/en.json
+++ b/src/renderer/i18n/en.json
@@ -407,7 +407,7 @@
     "firmwareSources": "Source code",
     "calloutTitle": "Important!",
     "factoryResetTitle": "Factory reset",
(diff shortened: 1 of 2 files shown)


src/renderer/i18n/en.json
```diff
@@ -406,6 +406,9 @@
     },
     "firmwareSources": "Source code",
     "calloutTitle": "Important!",
+    "factoryResetTitle": "Factory reset",
+    "factoryResetDescription": "Erase all settings layouts on your keyboard, restoring it to factory defaults.",
```
algernon, line 410:
```suggestion
    "factoryResetDescription": "Erase all customizations from your keyboard, restoring it to factory defaults.",
```

Not sure about the suggested wording either, but "all settings layouts" doesn't sound right to me.

src/renderer/i18n/en.json
```diff
@@ -406,6 +406,9 @@
...
+"factoryResetWarning": "Selecting this option will not backup your current layout or settings.",
```
algernon, line 411:
```suggestion
"factoryResetWarning": "Selecting this option will have the new firmware start up with factory defaults, no customizations applied.",
```

algernon (changes requested):
Left a few wording suggestions, going to take the PR for a test drive now. I'll report back how it went.

algernon:
Flashed my keyboard a few times with the factory reset enabled, and a few times with it disabled, worked as expected.

One note, to be fixed at some point later, perhaps:
Screenshot from 2022-09-28 00-47-14

With the added widgets, we no longer fit onto a single screen on my 1920x1080 window. Not the end of the world, but it feels like screen is getting a bit too... crowded?

A quick fix would be to move the stepper to the top, but that's outside of the scope of this PR. I'll open an issue about it once this is in.

src/renderer/i18n/en.json
obra: I want the scarier warning here.

algernon (approved):

PR #1110, opened 2022-09-28
make key labels more legible

### LINKED ISSUE #1094: increase font size in key labels
**Describe the bug**
In the Layout & Colormap Editor, key labels are small enough to be nearly illegible to me, maybe especially on a Retina (or maybe other HiDPI) display.

**To Reproduce**
1. Go to 'Layout & Colormap Editor'.
2. Try to read the key labels on the keyboard preview.

**Expected behavior**
Key labels should be more legible. Changing to `font-size: 10px` on both `.layer` and `.extra-key` by editing the CSS in Developer Tools seems to result in a good balance of increasing readability and avoiding overflowing the key outlines. A lot of emoji key labels are still difficult for me to read at this size, so maybe there needs to be an even larger size used for emoji key labels.

**Screenshots**
N/A; probably need to view on an actual screen to see.

 - OS: macOS 10.15.7 (19H2026)
 - Chrysalis Version: 0.11.4

**Additional context**
N/A

tlyu:
Increase font size for non-short key labels. This increases legibility.

Fixes #1094.

### DIFF
--- a/src/styles/keymap.css
+++ b/src/styles/keymap.css
@@ -21,7 +21,7 @@ button {
 
 .layer {
   font-family: "Source Code Pro", monospace;
-  font-size: 10px;
+  font-size: 9.5px;
   font-weight: 700;
   user-select: none;
 }
@@ -31,7 +31,7 @@ button {
 }
 
 .extra-key {
-  font-size: 10px;
+  font-size: 9.5px;
 }
 
 .key:hover {
--- a/src/styles/keymap.css
+++ b/src/styles/keymap.css
@@ -21,7 +21,7 @@ button {
 
 .layer {
   font-family: "Source Code Pro", monospace;
-  font-size: 8px;
+  font-size: 10px;
   font-weight: 700;
   user-select: none;
 }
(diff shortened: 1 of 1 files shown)


tlyu:
Please check with a variety of key labels to make sure that longer labels don't overflow the key shapes.

algernon:
For the record, even with the current size, we do have *some* labels that overflow some keys. If this font size increase causes some more to overflow, as long as they're not commonly used labels, we should merge it nevertheless.

I'll have a closer look after some sleep.

algernon:
Unfortunately, this makes `ShiftTo #1` overflow 1u keys on the Atreus:

Screenshot from 2022-09-29 21-24-56

If it were things like `S+C+A+Control/" overflowing, I'd be fine with that. But `ShiftTo #N`  overflowing is something I'd consider a blocker, I'm afraid.

tlyu:
Does 9.5px still seem to overflow on the Atreus? I think I don't have an easy way to preview it without the hardware. I might be willing to drop down to 9px, but that's starting to become almost as unreadable as 8px.

obra:
Could we use CSS to truncate at the shape boundary?

obra:
Key legend sizes and keyboard both size got tweaked since this was first proposed.

PR #1128, opened 2022-10-01
Improve the layout card print style

### LINKED ISSUE #948: Layout Cards : Inverse Atreus colors for printing (Black symbols on White board)
Printing layers is awesome and so useful to learn touch-typing, but using a full black cartridge is :wastebasket: 

The Atreus keyboard is so nice, and seems real, but as it is very dark, not really eco-friendly to print.

Having a white keyboard with black (with colors keys option) signs would save a lot of ink : 
Something that look more like the concrete laminated card.

noerw:
- Fix white text on white background of layer labels
- Don't print UI elements (toasts)
- Print uncolored keys in white to save ink (fixes #948)


noerw:
I now see that between 0.11.5 and current master the layout card style was modified to not show any color (but I don't find the commit..?).
I think with this PR that change could be reverted, so colored keys still are printed in color, but uncolored keys don't use unnecessary ink.

noerw:
> . The issue in https://github.com/keyboardio/Chrysalis/issues/948's case is that the background we use for the Atreus keyboard is a black keyboard.

Oh I see. I doubt this PR will fix it then, but I can't test easily as I don't have the hardware.

> It's actually a setting in Preferences, where you can toggle whether to show colors on the layout cards or not, added in https://github.com/keyboardio/Chrysalis/commit/7dc67917f61b82c2128de23b611c26a84710e49e:

Good to know, got it.

>  I'd keep printing uncolored keys as black if colors are turned on, because that matches how they're displayed in the editor.

imho that is a weak argument. Actually, going by that argument, the layout page shouldn't have a separate print style and should display what is printed (eg white background, black text, etcpp).
If you drop this the premise of printing black keys, there is no additional edge case; uncolored keys just would never be printed in black, irrespective of that or a new setting. So this PR should enable just that behaviour.
But I'd also be happy to remove the style that makes uncolored keys white, if you still prefer that.

Anyway, I just tested again and see that it does not work correctly with printing colors turned on, I'll have to revise this, give me a second.

noerw:
Ok, I have removed that change, so only the text color improvement remains.

noerw:
> > > . The issue in #948 case is that the background we use for the Atreus keyboard is a black keyboard.
> > 
> > 
> > Oh I see. I doubt this PR will fix it then, but I can't test easily as I don't have the hardware.
> 
> I just checked, and it doesn't, the Atreus background is still there, printed, in all its black glory. :(

There is a workaround in the print style to *ensure* backgrounds are printed; this was added in 9a187ef9a485cc5bc5ff444e60089bde7c1d1d91.
If this is the only background image in this context, I guess we can just remove that bit of CSS - or are there other background images we expect to print?

noerw:
Sorry I just made another change; added a print button, can you re-review? Then I'll squash the commits

noerw:
The i18n json file has no "ui" section, only `preferences.ui` (aside: note that indentation in all following lines (>294) is messed up), so I put it into a new top level section `layoutCards` analog to the other pages, hope thats' fine.

PR #1211, opened 2022-10-25
Restyle our "keyButton" to look more like our keyboard buttons and less like action buttons

obra:


obra:
That reduction is very, very intentional and was the main point of the rework. The old buttons were way, way too big, all out of proportion to the rest of the UI.

obra:
Among other things, this is the first step toward being able to move more of these buttons to the floating key picker panel. With the old style, we can fit very, very few of them on screen.

PR #1231, opened 2022-11-12
Support TopsyTurvy

davispw:
Adds "TopsyTurvy" modifier as alternative to "Shift".

### DIFF
--- a/src/api/focus/keymap/db/modifiers.js
+++ b/src/api/focus/keymap/db/modifiers.js
@@ -22,6 +22,7 @@ const modMap = {
   altgr: 1 << 10,
   shift: 1 << 11,
   gui: 1 << 12,
+  topsyturvy: 53293,
 };
 
 const addModifier = (keyCode, mod) => {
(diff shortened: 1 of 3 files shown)


davispw:
Requesting initial feedback.

Note, the TopsyTurvy kaleidoscope plugin doesn't currently respond to `onNameQuery`, so for now, disable "Hide features not supported by your keyboard's current firmware" to use.  That'll be a separate firmware PR.

lldata:
I tested it and it works great. 

But it took me a while to figure out that I had to add the TopsyTurvy plugin to the firmware manually for it to kick in.

This is probably obvious to you guys, but the plugin should probably be added to the default firmware, so users don't have to figure this out for them self.

lldata:
One more comment:

I know the plugin is called TopsyTurvy and I am been flashing my firmware for years, but would it make sense to show a more newbie friendly label in Chrysalis? 

How about "Invert Shift"

obra:
I'm so sorry this hung out for so long. It was at the top of @algernon's queue and then he got sick and hasn't been working for the past few months. I just merged it manually, along with @lldata's suggestion to call it "Invert shift" in the UI.

Thanks so much for the contribution. I'm going to try to test out your other PR later today.

PR #1235, opened 2022-11-19
Support OneShot Meta Keys

davispw:


obra:
Thank you! I've just merged this and then followed it up with another commit massaging the UI a bit. Again, I'm so sorry it languished so long.

Issue #1069, opened 2022-09-25
Add bépo keyboard layout


BenjaminGalliot:
For bépo users, it should be nice to have in Preferences the bépo layout to make our life easier!

I give you the official layout image of https://bepo.fr/wiki/Accueil

![Capture d’écran du 2022-09-25 12-46-29](https://user-images.githubusercontent.com/23100612/192139938-0a284fd5-8e51-4b89-9eb8-0e46fd011923.png)

It is not so easy, because some characters are tricky: `'` and `’` (apostrophes), ` ` and ` ` (spaces), so I also give you the X11/xkb/symbols/fr file part of interest:

```
// Version 1.1rc2 of the Bépo keyboard layout, 
// normalized by the AFNOR NF Z71‐300 norm.
// 
// Layout: https://bepo.fr/wiki/Version_1.1rc2
// Normalization: https://normalisation.afnor.org/actualites/faq-clavier-francais/
// 
// ┌────┬────┬────┬────┬────┬────┬────┬────┬────┬────┬────┬────┬────╔═════════╗
// │ # ¶│ 1 „│ 2 “│ 3 ”│ 4 ⩽│ 5 ⩾║ 6  │ 7 ¬│ 8 ¼│ 9 ½│ 0 ¾│ ° ′│ ` ″║         ║
// │ $ –│ " —│ « <│ » >│ ( [│ ) ]║ @ ^│ + ±│ - −│ / ÷│ * ×│ = ≠│ % ‰║ <--     ║
// ╔════╧══╗─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─┴──┬─╚══╦══════╣
// ║  |<-  ║ B _│ É  │ P §│ O Œ│ È `║ !  │ V  │ D  │ L £│ J  │ Z  │ W  ║   |  ║
// ║  ->|  ║ b |│ é ´│ p &│ o œ│ è `║ ˆ ¡│ v ˇ│ d ∞│ l /│ j  │ z ―│ w  ║ <-'  ║
// ╠═══════╩╗───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───┴┬───╚╗     ║
// ║        ║ A Æ│ U Ù│ I ˙│ E ¤│ ; ,║ C ©│ T ™│ S ſ│ R ®│ N  │ M  │ Ç ©║     ║
[...]

PR #1303, opened 2023-08-23
Add bépo layout as a host keyboard layout

nicos68:
This PR offers a solution to https://github.com/keyboardio/Chrysalis/issues/1069. See my comment [here](https://github.com/keyboardio/Chrysalis/issues/1069#issuecomment-1659769672). The idea is to add the bépo layout among the selectable layouts in the Layout editor section of the User Interface preferences:
image
As said in my comment the result is not perfect (so don’t hesitate to make suggestions) but it is a dramatic improvement for bépo users:
image
I only have a Fedora machine so I couldn’t test this on anything, so don’t hesitate to make more tests.

### DIFF
--- a/data/cldr/keyboards/windows/fr-bépo-t-k0-windows.xml
+++ b/data/cldr/keyboards/windows/fr-bépo-t-k0-windows.xml
@@ -1,0 +1,290 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE keyboard SYSTEM "../dtd/ldmlKeyboard.dtd">
+<keyboard locale="fr-t-k0-windows">
+	<version platform="10" number="$Revision$"/>
+	<names>
+		<name value="French bépo"/>
+	</names>
+	<settings fallback="omit" transformPartial="hide"/>
+	<keyMap>
+		<map iso="E00" to="$"/> <!-- ` -->
+		<map iso="E01" to="\u{22}"/> <!-- 1 -->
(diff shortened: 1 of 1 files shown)


obra:
So the other custom layouts assume a "traditional" host keymap. (Dvorak and Colemak both assume your host's layout is set to QWERTY.)   Among other things, that allows the keyboard to work correctly when used on an "unconfigured" keyboard.

Would it make sense to have the Bépo layout assume an AZERTY configured computer instead?

nicos68:
> So the other custom layouts assume a "traditional" host keymap. (Dvorak and Colemak both assume your host's layout is set to QWERTY.) Among other things, that allows the keyboard to work correctly when used on an "unconfigured" keyboard.
> 
> Would it make sense to have the Bépo layout assume an AZERTY configured computer instead?

Hi,
on my machine the host os keyboard layout is set to bépo. 
image
This makes the assumption that the user has set their host os keyboard layout to bépo, which is natively available on linux. So if I understand you question correctly, I would say the answer is no.

obra:
I understand that for your particular setup, a native Bépo configuration is absolutely the right thing, but need to think about the more general case of a default configuration for users who want Bépo to "just work" for their keyboard.   The right answer might be to offer *both* configurations.

nicos68:
Maybe there could be a "French bépo (native)" and a "French bépo (emulated)" entry with tooltips to explain what each is.

PR #1379, opened 2025-04-11
Fix #1378: correct TopsyTurvy keycode range

davispw:

Issue #1378, opened 2025-04-11
TopsyTurvey key codes are wrong


davispw:
**Describe the bug**

TopsyTurvy / "Invert Shift" is broken.

**To Reproduce**

1. Select the "A" key
2. In the Modifiers tab, enable "Invert Shift"
3. Save settings to keyboard
4. Type: "A" (**actual result**: _nothing happens - key is broken_)
5. Go to the "Advanced" tab and note the key code (**actual result**: _32775_)
6. Change the key code to 53297 (**actual result**: _"#53297" is shown instead of "Ƨ+A"_)
7. Save settings to the keyboard, type "A" (**actual result**: _capital A is input_)

**Expected behavior**

Key code for inverted "A" should be `53297`.

**Additional context**

https://github.com/keyboardio/Chrysalis/commit/a0128dd506d293ee02339e91bd0c7fe05787d85c changed the TopsyTurvy keycode range start from "`53293`" to "`(1 << 15) | (1 << 0) | (1 << 1)`".  It doesn't seem to match the one derived from https://github.com/keyboardio/Kaleidoscope/blame/1221ac08226d314b0aea622b64ad7211b60d6184/plugins/Kaleidoscope-Ranges/src/Kaleidoscope-Ranges.h#L80.

PR #1380, opened 2025-04-12
TopsyTurvy cannot be combined with other mods

davispw:
This PR prevents a collision with DynamicMacros key range.  For example, DM#0 has keycode 53596.  Without this PR, this displays as "C+Ƨ+{" and the macro can't be edited.

This PR:
1. Prevents generating modifier combinations that include `topsyturvy` with other mods.  This avoids the display issue where DynamicMacros \#0-9 decoded to TopsyTurvy+Control.
2. In the Modifiers tab, disables the TopsyTurvy radio button when any other mod is selected, and vice versa.  This prevents the user creating invalid combinations.

This is a quick follow-up to PR #1379, addressing issue #1378.

### DIFF
--- a/src/api/focus/keymap/db/modifiers.js
+++ b/src/api/focus/keymap/db/modifiers.js
@@ -67,7 +67,7 @@ const setModifiersLabel = (key, mods) => {
 };
 
 const generateModifierCombinations = () => {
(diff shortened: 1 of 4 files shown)


davispw:
Looks like the lint errors are on main: https://github.com/keyboardio/Chrysalis/actions/runs/14414362914.

PR #1394, opened 2025-07-08
Add backup/restore support for EEPROM settings

thehenkan:
This adds a button to dump EEPROM settings to a JSON file, and a button to restore them by uploading a backup.

### DIFF
--- a/src/api/focus.js
+++ b/src/api/focus.js
@@ -549,8 +549,7 @@ class Focus {
     "typingbreaks.rightMaxKeys",
   ];
 
-  eepromBackupCommands = [
-    ...this.eepromRestoreCommands,
+  eepromBackupOnlyCommands = [
     "help",
     "version",
     "plugins",
@@ -558,6 +557,11 @@ class Focus {
     "settings.valid?",
     "settings.version",
     "settings.crc",
+  ];
+
+  eepromBackupCommands = [
+    ...this.eepromRestoreCommands,
+    ...this.eepromBackupOnlyCommands,
   ];
   async readKeyboardConfiguration() {
     const backup = {};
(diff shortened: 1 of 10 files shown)


obra:
I was starting to look at this today. I think my assumption had been that
it would make more sense in the import/export ui where we do keymaps. For
most users, I suspect they would expect keymaps and config to be bundled
together. (and the existing export field  should be exporting the config)

What do you think?

thehenkan:
Yeah that makes sense to me. Do you mean as in the existing backup/restore buttons for the keyboard layout should just be extended to encompass all EEPROM settings instead of just the keymap, or simply moving the buttons I added to that screen?

obra:
So, my recollection is that the existing backup button *already* includes the eeprom settings. We just never handled them on import.  Which makes me feel like adding a checkbox for the restore button to "overwrite keyboard settings" would make more sense than new buttons. And I think that it *might* make sense to have that button checked by default.  This would mean that restore from an automated backup that gets downloaded during a firmware flash would become easy and straightforward. Which would be amazing.

thehenkan:
It's true that the backup button already includes the eeprom settings. I've been playing around a bit with adding that to the layout restore button just now. One thing I noticed was that the layout restore loads the layout into the UI for inspection, and then has a "Save changes" button for actually writing to the device, whereas the functionality I added to the settings page directly wrote the settings to the device and then updated the settings UI to reflect the new state.

We could make the full eeprom restore also do a separate upload & save stages, but the current layout backup screen doesn't offer a way to visualise the changes. You can't even go to the settings screen to see them without discarding the unsaved changes. Conversely, having the button sometimes write directly and sometimes not would of course be very unintuitive behaviour.

So far I've thought of merging the "My Keyboard" settings page into the layout backup page, or alternatively adding a list of settings that were changed (and their previous and new states) but without an option to change them. What do you think?

thehenkan:
I moved it to the existing restore button now. The behaviour of the restore can be set to only restore layout via a drop down menu. I also added a basic table showing old and new values for loaded settings, but only for the changed settings. Let me know what you think.

obra:
Thanks! Sorry it took a minute to get this merged

